PR #5, opened 2021-05-18
Java Web Practice Done

mankokolya:


src/main/java/mate/lib/exception/DbConnectionException.java
```diff
@@ -0,0 +1,7 @@
+package mate.lib.exception;
+
+public class DbConnectionException extends RuntimeException {
+    public DbConnectionException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
```
Dalai-S, line 7: i dont see the reason for this exceptions now

src/main/java/mate/model/Manufacturer.java
```diff
@@ -7,6 +7,9 @@
     private String name;
     private String country;
 
+    public Manufacturer() {
+    }
```
Dalai-S, line 11: why do you need an empty constructors?

src/main/webapp/WEB-INF/views/cars/getAll.jsp
```diff
@@ -0,0 +1,33 @@
...
+  <c:forEach items="${cars}" var="car">
+      <tr>
+          <td><c:out value="${car.id}"></c:out></td>
+          <td><c:out value="${car.model}"></c:out></td>
+          <td><c:out value="${car.getManufacturer().name}"></c:out></td>
+          <td><c:out value="${car.getManufacturer().country}"></c:out></td>
+          <c:forEach items="${car.getDrivers()}" var="driver">
+              <td><c:out value="${driver.name}"></c:out></td>
+              <td><c:out value="${driver.licenseNumber}"></c:out></td>
+          </c:forEach>
```
Dalai-S, line 27: if really i dont like the way how it would looks like on the screen. if there are many of drivers, you will have a two columns and long long list of names and id

Dalai-S (approved):

src/main/java/mate/lib/exception/PropertyFileReadingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.lib.exception;
+
+public class PropertyFileReadingException extends RuntimeException {
```
RomanDubovskyi, line 3: I don't see, where this exception is being used, same with DriverLoadingException

RomanDubovskyi (approved):

src/main/java/mate/lib/exception/PropertyFileReadingException.java
mankokolya: I read properties from properties file in previous project and tried to use it in this one but got exception "can't read properties file". That's why I moved everything to util class but forget to delete custom exceptions

PR #14, opened 2021-05-23
first solution

dvdovynets:


src/main/java/mate/controller/car/GetAllCarsController.java
```diff
@@ -0,0 +1,33 @@
...
+
+@WebServlet(urlPatterns = "/cars/all")
+public class GetAllCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
```
freedmanMain, line 24: Remove unnecessary services.

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add")
```
freedmanMain, line 16: It is not necessary to specify "urlPatterns" explicitly.

freedmanMain (approved):

src/main/java/mate/controller/car/GetAllCarsController.java
dvdovynets: Cleared. My fault

src/main/java/mate/controller/car/AddCarController.java
dvdovynets: It wouldn't want to work other way

Resci (approved):
Good job!

PR #16, opened 2021-05-24
Web Practice Solution

vuyko23:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet (urlPatterns = "/car/driver/add")
```
SerKisJD, line 15: due to checklist: cars/drivers/add

SerKisJD (approved):

src/main/java/mate/controller/AddDriverToCarController.java
MaliukDaria: I agree, let's make all urls uniformed

MaliukDaria (changes requested):

MaliukDaria (approved):
Good Job!

PR #25, opened 2021-05-24
First solution

Ihor-Nykyforov:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Long carId = Long.valueOf(req.getParameter("car"));
+        Long driverId = Long.valueOf(req.getParameter("driver"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
```
Reastador, line 41: you can write carService.addDriverToCar(driverService.get(driverId), carService.get(carId));
in yhis case you will don't create 2 objects and use 2 instances created in class header

Reastador (approved):

src/main/java/mate/controller/AddDriverToCarController.java
MaliukDaria: Both options are suitable

MaliukDaria (approved):
Good job!)

PR #89, opened 2021-10-11
1.0 Solution "web-taxi"

Chupiak:


pom.xml
```diff
@@ -78,4 +84,4 @@
             </plugins>
         </pluginManagement>
     </build>
-</project>
+</project>
```
RomanDubovskyi, line 87: missing an empty line at the end

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,51 @@
...
+public class AddDriverToCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        try {
```
RomanDubovskyi, line 26: it's not quite clear what code here, generates exception that you want to catch and redirect user to error page. Let's make try catch block smaller + catching RuntimeException is a bad practice, if you have bad code here, for example NPE or IOB, user will see error page, but it's not his fault. Ideal solution would be custom exception that is related only with invalid entered data. But for this task it would be to complicated so I suggest we just move unrelated code from try block

src/main/webapp/WEB-INF/views/manufacturers/manufacturer_created.jsp
```diff
@@ -0,0 +1,19 @@
...
+      text-align: center;
+    }
+  </style>
+  <title>Manufacturer was added</title>
+</head>
+<body>
+<h1>Manufacturer ${manufacturer} was successfully registered!</h1>
+<h2><a href="${pageContext.request.contextPath}/index">Back to main menu</a></h2>
+</body>
+</html>
```
RomanDubovskyi, line 19: check jsp files for empty lines at the end 
P.S.
you can configure Intellij IDEA to do that automatically

RomanDubovskyi (changes requested):

src/main/webapp/WEB-INF/views/manufacturers/manufacturer_created.jsp
Chupiak: ok

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,51 @@
...
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        try {
+            Car car = carService.get(carId);
```
RomanDubovskyi, line 31: Ideally with your logic I would then return Optional from get method on the service layer, and here customly redirect user for error page

PR #90, opened 2021-10-11
Implemented controllers

VladyslavLekhter:


Yelinskyi (approved):

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        Car car = new Car(req.getParameter("model"),
+                manufacturerService.get(Long.parseLong(req.getParameter("manufacturerId"))));
```
MaliukDaria, line 31:
```suggestion
        String model = req.getParameter("model");
        Long manufacturerId = Long.parseLong(req.getParameter("manufacturerId"));
        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
        Car car = new Car(model, manufacturer);
```

I think it's more readable this way

PR #92, opened 2021-10-11
HW22 web practice solution

Yehor-Zapara:


src/main/webapp/WEB-INF/views/cars/all.jsp
```diff
@@ -0,0 +1,49 @@
...
+        <td>DELETE</td>
+    </tr>
+    <c:forEach items="${cars}" var="car">
+        <tr>
+            <td><c:out value="${car.id}" /></td>
+            <td><c:out value="${car.model}" /></td>
+            <td><c:out value="${car.manufacturer.id}" /></td>
+            <td>
+                <c:forEach items="${car.drivers}" var="driver">
+                    <c:out value="${driver.id} " />
```
Koriastras, line 37: I think it would be better to show more information about drivers and manufacturers (but it is only a suggestion)

src/main/webapp/WEB-INF/views/drivers/all.jsp
```diff
@@ -0,0 +1,43 @@
...
+        border-collapse: collapse;
+        text-align: center;
+    }
+    td {
+        padding: 5px;
+    }
+    tr:nth-child(even) {
+        background-color: #DDDDDD;
+    }
+</style>
```
Koriastras, line 21: you can create a file  "styles.css" and use here , for example <style><%@include file="/WEB-INF/styles.css"%></style>

src/main/webapp/WEB-INF/tablesStyle.css
```diff
@@ -0,0 +1,11 @@
...
+    border: 1px solid black;
+    border-collapse: collapse;
+    text-align: center;
+}
+td {
+    padding: 5px;
+}
+tr:nth-child(even) {
+    background-color: #DDDDDD;
+}
```
Koriastras, line 11: empty line

Koriastras (approved):

src/main/webapp/web.xml
```diff
@@ -11,4 +11,8 @@
         <servlet-name>index</servlet-name>
         <url-pattern>/index</url-pattern>
     </servlet-mapping>
+    <servlet-mapping>
+        <servlet-name>index</servlet-name>
+        <url-pattern>/</url-pattern>
+    </servlet-mapping>
```
MaliukDaria, line 17: If you use @WebServlet annotation, it's better to use it everywhere(better to add it to IndexController too)

PR #106, opened 2021-10-13
Created controllers and jsp pages for the application

Zbarazkyy:


src/main/webapp/WEB-INF/views/service/createCar.jsp
```diff
@@ -0,0 +1,20 @@
...
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>add car</title>
+</head>
+<body>
+<h1>Enter the model and select the manufacturer from the list</h1>
+<br>
+<form method="post" action="${pageContext.request.contextPath}/car/add">
+    Model <input type="text" name="carModel"/><br>
```
Max-Pochepets, line 11: what if we won't input anything here?

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -34,7 +34,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
         List<Car> cars = carService.getAll();
         req.setAttribute("drivers", drivers);
         req.setAttribute("cars", cars);
-        req.getRequestDispatcher("/WEB-INF/views/service/addDriverToCar.jsp").forward(req, resp);
+        req.getRequestDispatcher("/WEB-INF/views/service/carsDriversAdd.jsp").forward(req, resp);
```
Max-Pochepets, line 37:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/cars/drivers/add.jsp").forward(req, resp);
```
same with all others

PR #114, opened 2021-10-13
created all controllers, implemented doGet, doPost methods and created all jsp files

sich-mykhailo:


src/main/java/mate/controller/CreateManufacturerController.java
```diff
@@ -0,0 +1,32 @@
...
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/manufacturers/create.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) {
+        String name = req.getParameter("name");
+        String country = req.getParameter("country");
+        Manufacturer manufacturer = new Manufacturer(name, country);
+        manufacturerService.create(manufacturer);
```
MaliukDaria, line 30: lets redirect user for some page after we add manufacturer(fix in all places)

MaliukDaria (changes requested):

PR #128, opened 2021-11-10
added solution

YaroslavYarmoshyk:


src/main/java/mate/controller/AssignmentOfDriversToCarsController.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars_drivers")
+public class AssignmentOfDriversToCarsController extends HttpServlet {
```
MaliukDaria, line 16:
```suggestion
public class AssignDriversToCarsController extends HttpServlet {
```

src/main/java/mate/controller/AssignmentOfDriversToCarsController.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars_drivers")
```
MaliukDaria, line 15:
```suggestion
@WebServlet(urlPatterns = "/cars/drivers/add")
```

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,44 @@
...
+            throws ServletException, IOException {
+        String carModel = req.getParameter("model");
+        String makerName = req.getParameter("maker");
+        String makerCountry = req.getParameter("maker_country");
+        Manufacturer maker = manufacturerService.getAll().stream()
+                .filter(m -> m.getName().equals(makerName)
+                && m.getCountry().equals(makerCountry))
+                .findFirst()
+                .orElseThrow(() -> new RuntimeException("Can't get maker "
+                        + makerName + " from DB"));
```
MaliukDaria, line 33: It's bad practice to get all entities from db, and then filter them. What if you have 100 000 makers? If you want to get manufacturer by name, you should create separate method in dao and service layers. But I think this is initially not a good idea, because theoretically the names of two different manufacturers can be the same

src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,45 @@
...
+        Driver driver = new Driver(driverName,licenseNumber);
+        for (Driver tempDriver : drivers) {
+            if (tempDriver.getName().equals(driver.getName())
+                    && tempDriver.getLicenseNumber()
+                    .equals(driver.getLicenseNumber())) {
+                throw new RuntimeException("Driver " + driver.getName()
+                        + " with licenseNumber " + driver.getLicenseNumber()
+                        + " is already exist");
+            }
+        }
```
MaliukDaria, line 35: Same here. First of all we definetelly could have 2 drivers with the same name. Second - we could check the uniqueness of the license number, but we should create separate methods in dao and service layers to find driver by license number(or check if there is such license number in the db). And I think we should do this check on service layer, not in controller, as this is a part of business logic. (Same in all other places in the code)

src/main/java/mate/controller/CreateMakerController.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Manufacturer;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/makers/create")
```
MaliukDaria, line 14:
```suggestion
@WebServlet(urlPatterns = "/manufacturers/create")
```

src/main/java/mate/controller/GetAllCarController.java
```diff
@@ -0,0 +1,27 @@
...
+@WebServlet(urlPatterns = "/cars")
+public class GetAllCarController extends HttpServlet {
+    private static Injector injector = Injector.getInstance("mate");
+    private static CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Car> allCars = carService.getAll();
+
```
MaliukDaria, line 23: redundant empty line

src/main/java/mate/controller/GetAllDriverByCarIdController.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/driversByCar")
```
MaliukDaria, line 15:
```suggestion
@WebServlet(urlPatterns = "/cars/drivers")
```

look to the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-web/web-practice/java-web-practice) there is some suggestions about url naming

src/main/java/mate/dao/DriverDaoImpl.java
```diff
@@ -109,4 +131,13 @@ private Driver getDriver(ResultSet resultSet) throws SQLException {
...
     }
+
+    private Driver parseDriverFromResultSet(ResultSet resultSet) throws SQLException {
+        long driverId = resultSet.getObject("id", Long.class);
+        String driverName = resultSet.getNString("name");
+        String licenseNumber = resultSet.getNString("license_number");
+        Driver driver = new Driver(driverName, licenseNumber);
+        driver.setId(driverId);
+        return driver;
+    }
```
MaliukDaria, line 142: why do we need this method? We already have `getDriver(ResultSet resultSet)` with the same logic

src/main/java/mate/lib/Injector.java
```diff
@@ -164,6 +164,13 @@ private void setValueToField(Field field, Object instanceOfClass, Object classTo
                     classes.addAll(findClasses(file, packageName + "."
                             + file.getName()));
                 } else if (file.getName().endsWith(".class")) {
+
+                    /*String controller = packageName + '.'
+                            + file.getName().substring(0, file.getName().length() - 6);
+                    if (controller.contains("Controller")) {
+                        continue;
+                    }*/
+
```
MaliukDaria, line 173: lets remove comments from code

MaliukDaria (changes requested):

src/main/java/mate/controller/CreateMakerController.java
```diff
@@ -0,0 +1,36 @@
...
+    private static Injector injector = Injector.getInstance("mate");
+    private static ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String makerName = req.getParameter("maker_name");
+        String country = req.getParameter("country");
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
```
MaliukDaria, line 25: where do we use this?

MaliukDaria (changes requested):

RomanDubovskyi (approved):

PR #129, opened 2021-12-03
solution done

ikhominitch:


src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,42 @@
...
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Manufacturer> allManufacturersList = manufacturerService.getAll();
+        req.setAttribute("maxId", allManufacturersList.stream()
+                .mapToLong(Manufacturer::getId)
+                .max()
+                .orElse(1L));
```
MaliukDaria, line 28: It's really bad practice to get all records from db and then filter them. If you want to get the maximum index from a table in the database, you need to create a separate query to the database. But initially I do not see how this information can be used, you may not have records with a specific index in the database so the maximum index is not particularly informative

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,49 @@
...
+        List<Driver> allDriversList = driverService.getAll();
+        List<Car> allCarsList = carService.getAll();
+        req.setAttribute("maxDriverId", allDriversList.stream()
+                .mapToLong(Driver::getId)
+                .max()
+                .orElse(1L));
+        req.setAttribute("maxCarId", allCarsList.stream()
+                .mapToLong(Car::getId)
+                .max()
+                .orElse(1L));
```
MaliukDaria, line 33: same here

MaliukDaria (changes requested):

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,38 @@
...
+public class AddCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Manufacturer> allManufacturersList = manufacturerService.getAll();
```
MaliukDaria, line 24: where do we use allManufacturersList?

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,41 @@
...
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Driver> allDriversList = driverService.getAll();
+        List<Car> allCarsList = carService.getAll();
+        req.getRequestDispatcher("/WEB-INF/views/cars/drivers/add.jsp").forward(req, resp);
```
MaliukDaria, line 26: where do we use allDriversLis and allCarsList?

MaliukDaria (changes requested):

MaliukDaria (approved):
Good Job! 🔥

PR #190, opened 2022-03-19
Implementhed all neeeded controllers and created 5 jsp pages

DaniilOpryshko:
List of pages: 
- Cars page (List of all cars, here u can delete, create or update car also it has link on car drivers page)
- Car drivers (contains list of certain car drivers and u able to delete or add driver to car)
- Manufacturers page (list of all manufacturers, delete, update, add manufacturer)
- Drivers page (list of all drivers, delete, update, add driver)
- Main page (Welcome message and nav bar to other pages)

### DIFF
--- a/src/main/java/mate/controller/cars/AddCarController.java
+++ b/src/main/java/mate/controller/cars/AddCarController.java
@@ -1,0 +1,36 @@
+package mate.controller.cars;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add", loadOnStartup = 1)
+public class AddCarController extends HttpServlet {
+    private static final Injector injector
+            = Injector.getInstance("mate");
+    private static final CarService carService
+            = (CarService) injector.getInstance(CarService.class);
+    private static final ManufacturerService manufacturerService
+            = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        String model = req.getParameter("carModel");
+        car.setModel(model);
+        car.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturerId"))));
+        if (!model.isEmpty()) {
+            carService.create(car);
+        }
+        resp.sendRedirect(req.getContextPath() + "/cars");
+    }
+}
--- a/src/main/java/mate/controller/cars/CarDriversController.java
+++ b/src/main/java/mate/controller/cars/CarDriversController.java
@@ -1,0 +1,36 @@
+package mate.controller.cars;
+
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.ServletException;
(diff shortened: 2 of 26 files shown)


src/main/java/mate/controller/cars/CarDriversController.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars/drivers")
+public class CarDriversController extends HttpServlet {
```
nastyaprokhorenko, line 17: we should start controller’s name with a verb
```suggestion
public class GetCarDriversController extends HttpServlet {
```

src/main/webapp/WEB-INF/views/cars/carDrivers.jsp
```diff
@@ -0,0 +1,68 @@
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>Driver Control Panel</title>
+    <style>
+        table, th, td {
```
nastyaprokhorenko, line 7: for styles it is better to use a separate css file. but it's not a mistake in this task

nastyaprokhorenko (changes requested):

nastyaprokhorenko (approved):
looks good

src/main/java/mate/controller/cars/AddCarController.java
```diff
@@ -0,0 +1,36 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        String model = req.getParameter("carModel");
+        car.setModel(model);
+        car.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturerId"))));
+        if (!model.isEmpty()) {
+            carService.create(car);
+        }
```
KostyaNikiforov, line 33: Do we really need this checking? If so, I think it is better to move validation at the begin of the method, before you have created an object

src/main/webapp/WEB-INF/views/cars/cars.jsp
```diff
@@ -0,0 +1,87 @@
...
+            <td><c:out value="${car.getModel()}"/></td>
+            <td><c:out value="${car.getManufacturer().getName()}"/></td>
+            <td><a href="${pageContext.request.contextPath}/cars/drivers?carId=${car.id}"><c:out value="${car.drivers.size()}"/> drivers</a></td>
+            <td><a href="${pageContext.request.contextPath}/cars/delete?carId=${car.id}">DELETE CAR!</a> </td>
+        </tr>
+    </c:forEach>
+</table>
+
+<h2>Add Car</h2>
+<form name="form1" method="post" action="${pageContext.request.contextPath}/cars/add">
```
KostyaNikiforov, line 62: Let's come up with some more informative names (form1 - is not the best one)

KostyaNikiforov (changes requested):

src/main/java/mate/controller/cars/AddCarController.java
DaniilOpryshko: I added this check to prevent adding car without model. But now i think that it will be better to move this check to service layer. Am i right?

KostyaNikiforov: I don't think that any kind of validation is necessary in this task at all. But if you really would like to check if some input values is null, it is better to do at the begin of the method before creating an object and send to the user a corresponding message. Now it is not so important, anyway in a future you will use specific libraries for this

KostyaNikiforov (approved):

PR #191, opened 2022-03-22
Develop

OleksiyTitov:


LinkClink (changes requested):
Взгляни на все resp.sendRedirect()

LinkClink (changes requested):
.

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturerId"));
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        Car car = new Car();
+        car.setManufacturer(manufacturer);
+        car.setModel(model);
+        car.setDrivers(new ArrayList<>());
```
KostyaNikiforov, line 42: Passing an empty list is unnecessary because nothing will change in the database anyway.
And if you need an empty list, it is better to use this method:
`Collection.emptyList()`

KostyaNikiforov (approved):

PR #193, opened 2022-03-24
web-practice solution

sltnsrh:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        carService.addDriverToCar(driverService.get(driverId),
+                carService.get(carId));
+        resp.sendRedirect("/cars/drivers/add");
```
OleksiyTitov, line 32: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work. 
```suggestion
        resp.sendRedirect(req.getContextPath() + "/cars/drivers/add");
```
Review all other controllers and fix this error in them as well.

OleksiyTitov (changes requested):

src/main/java/mate/controller/GetAllDriversController.java
```diff
@@ -0,0 +1,25 @@
...
+public class GetAllDriversController extends HttpServlet {
+    private final Injector injector = Injector.getInstance("mate");
+    private final DriverService driverService = (DriverService) injector
+            .getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Driver> allDrivers = driverService.getAll();
+        req.setAttribute("drivers", allDrivers);
```
KostyaNikiforov, line 22: Let's make the naming of lists more consistent, you already called the list of cars in this way
```suggestion
        req.setAttribute("drivers", driverList);
```

src/main/java/mate/controller/GetAllCarsController.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class GetAllCarsController extends HttpServlet {
+    private final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector
+            .getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Car> carList = carService.getAll();
```
KostyaNikiforov, line 21: Or `allCars` here

KostyaNikiforov (approved):

PR #195, opened 2022-03-29
all needed controllers implemented (without advanced pack)

DzhuzhaMichael:


src/main/webapp/WEB-INF/views/drivers/all.jsp
```diff
@@ -0,0 +1,24 @@
...
+    <c:forEach items="${drivers}" var="driver">
+        <tr>
+            <td><c:out value="${driver.id}" /></td>
+            <td><c:out value="${driver.name}" /></td>
+            <td><c:out value="${driver.licenseNumber}" /></td>
+        </tr>
+    </c:forEach>
+</table>
+</body>
+</html>
```
NazariiYatsiuk, line 24: image

NazariiYatsiuk: image

NazariiYatsiuk (changes requested):
I see no disadvantages, just add a newline in the end of "/drivers/all.jsp" file

NazariiYatsiuk (approved):

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Manufacturer manufacturer = manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturer_id")));
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
KostyaNikiforov, line 39: Do we really need to pass an empty array here? What will change in a base? If so, it is better to use method Collections.emptyLIst()

KostyaNikiforov (approved):

PR #197, opened 2022-03-29
Workbranch

NazariiYatsiuk:


src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/create")
```
DzhuzhaMichael, line 16: Optional comment.
As far as I understood, according to the checklist we should use one style for creating URLs applying ".../add" for objects creation

DzhuzhaMichael (approved):
Looks good for me

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car car = new Car();
+        Manufacturer manufacturer = manufacturerService.get(Long.valueOf(
+                req.getParameter("manufacturer_id")));
+        car.setModel(req.getParameter("model"));
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
KostyaNikiforov, line 38: Do we really need to pass an empty array here? What will change in a base? If so, it is better to use method `Collections.emptyLIst()`

KostyaNikiforov (approved):

PR #201, opened 2022-03-30
Task done and checked. Codestyle checked also.

lugburz-hr:


src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,49 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        String manufacturerName = req.getParameter("manufacturer_name");
+        Manufacturer manufacturer = getManufacturerByName(manufacturerName).get();
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        carService.create(car);
+    }
```
o240789, line 38: We can redirect here to index.jsp or somewhere else. 
```suggestion
        carService.create(car);
        resp.sendRedirect(url);
    }
```

o240789 (approved):
We can redirect to other URL after some actions

src/main/java/mate/controller/AddCarController.java
KostyaNikiforov: I agree, let's redirect user to the page with all cars to make him see the result

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,49 @@
...
+    private Optional<Manufacturer> getManufacturerByName(String name) {
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        for (Manufacturer manufacturer: manufacturers) {
+            if (manufacturer.getName().equals(name)) {
+                return Optional.ofNullable(manufacturer);
+            }
+        }
+        return Optional.empty();
+    }
+}
```
KostyaNikiforov, line 49: This logic does not belong here. It's bad practice to filter objects with java but not SQl.

KostyaNikiforov: You need to get an entity by name - create the method on a dao layer

src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String name = req.getParameter("name");
+        String licenseNumber = req.getParameter("licenseNumber");
+        Driver driver = new Driver();
+        driver.setName(name);
+        driver.setLicenseNumber(licenseNumber);
+        driverService.create(driver);
```
KostyaNikiforov, line 31: Same here about redirection

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private Optional<Car> getCarByModel(String name) {
+        List<Car> cars = carService.getAll();
+        for (Car car: cars) {
+            if (car.getModel().equals(name)) {
+                return Optional.ofNullable(car);
+            }
+        }
+        return Optional.empty();
+    }
```
KostyaNikiforov, line 56: Same here about inappropriate logic

KostyaNikiforov (changes requested):

PR #213, opened 2022-04-15
Solution of the task

k1llzers:
Добрый день. Во время решения я столкнулся с такой проблемой .Когда я перенаправляю на другую страницу , когда идёт запрос на сервер часть адреса попадает в путь к страничке . Возможно подскажете как это исправить

)

### DIFF
--- a/src/main/java/mate/controller/AddDriverToCarController.java
+++ b/src/main/java/mate/controller/AddDriverToCarController.java
@@ -1,0 +1,38 @@
+package mate.controller;
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/addDriverToCar")
+public class AddDriverToCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService
+            = (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("WEB-INF/views/addDriverToCar.jsp").forward(req,resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driverId"));
+        Long carId = Long.valueOf(req.getParameter("carId"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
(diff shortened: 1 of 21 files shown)


MaliukDaria (changes requested):
dont understand your question

k1llzers:
Thank you , but now everything is good , I solved this problems)

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driverId"));
+        Long carId = Long.valueOf(req.getParameter("carId"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver,car);
+        resp.sendRedirect("/menu");
```
MaliukDaria, line 36: checklist: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.
Fix in all places

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/create/car")
```
MaliukDaria, line 17: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-web/web-practice/java-web-practice): Let’s use one style for creating URLs. Try putting at the first position in URL’s name the name of the model with which we work.
If we work with the manufacturer - let’s put it here: /[manufacturers]/…/… It is good practice to use plurals.

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturerId"));
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<Driver>());
```
MaliukDaria, line 33: I dont think we need this `car.setDrivers(new ArrayList<Driver>());`

MaliukDaria (changes requested):

src/main/java/mate/controller/CreateCarController.java
k1llzers: I think that we need this line because in CarDaoImpl when we create a new car we need List of drivers.

MaliukDaria: Lets create empty ArrayList in Cat constructor

MaliukDaria: Lets create empty ArrayList in Cat constructor

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #221, opened 2022-05-15
task web-practice

MarkShapka:
- implemented controllers
- added configuration web.xml
- added jsp pages


src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setId(Long.valueOf(req.getParameter("manufacturer_id")));
+        car.setManufacturer(manufacturer);
+        List<Driver> drivers = new ArrayList<>();
```
MaliukDaria, line 39: we don't need to set empty ArrayList here (we can make it Car default constructor)

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,44 @@
...
+        car.setDrivers(drivers);
+        carService.create(car);
+        resp.sendRedirect("/cars/add");
```
MaliukDaria, line 42: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-web/web-practice/java-web-practice): If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.
Fix an all places

MaliukDaria (changes requested):

PR #226, opened 2022-05-19
implemented solution jv-web-practice

OLEH-VIESHAIEV:


src/main/java/mate/model/Driver.java
```diff
@@ -49,4 +49,13 @@ public boolean equals(Object o) {
...
     }
+
+    @Override
+    public String toString() {
+        return "Driver{"
+                + "id=" + id
+                + ", name='" + name + '\''
+                + ", licenseNumber='" + licenseNumber + '\''
+                + '}';
+    }
```
nvak, line 60: you don't use this in project. We can delete it

Sofasmile (approved):

PR #235, opened 2022-05-20
Implemented JSP pages and their corresponding controllers

Bukozdem:


src/main/java/mate/controller/CreateCarsController.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setId(Long.valueOf(req.getParameter("manufacturer_id")));
+        car.setManufacturer(manufacturer);
+        car.setDrivers(Collections.emptyList());
```
MaliukDaria, line 34: I dont think we neet to set emptyList here. If we need new Car to be created with empty drivers list, lets make it in a Car constructor

src/main/java/mate/controller/GetAllCarsController.java
```diff
@@ -0,0 +1,30 @@
...
+    private CarService carService = (CarService) injector.getInstance(CarService.class);
+    private ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Car> cars = carService.getAll();
+        req.setAttribute("cars", cars);
+        req.setAttribute("manufacturers", manufacturerService.getAll());
```
MaliukDaria, line 27: why do we need manufacturers list here?

MaliukDaria (changes requested):

src/main/java/mate/controller/GetAllCarsController.java
Bukozdem: sorry, i forgot to delete this line

Nick97-git (approved):

PR #248, opened 2022-05-29
jv_homework_web-practice

etezaadi:
Created a new controller: creating a driver;
Created a new controller: getting all drivers;
Created a new controller: creating a manufacturer;
Created a new controller: creating a car;
Created a new controller: adding driver to car;
Created a new controller: Display all manufacturers with ability to delete a certain manufacturer;
Getting all drivers with ability to delete a certain driver;
Created a new controller: Display all cars.


src/main/java/mate/controller/car/CreateCarController.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturer_id"));
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
MaliukDaria, line 39: I dont think we need to set empty ArrayList here. We can set new ArrayList<>() to drivers in Car constructor

MaliukDaria (approved):

PR #264, opened 2022-07-04
Hw jv web practice

zim10093:


src/main/java/mate/controller/car/AllDriverFromCarController.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars/drivers")
+public class AllDriverFromCarController extends HttpServlet {
```
antonEEmus, line 21:
```suggestion
public class AllDriversFromCarController extends HttpServlet {
```

antonEEmus: pay attention to the namings

src/main/java/mate/controller/car/AllDriverFromCarController.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final int INDEX_OF_FIRST_PARAM = 0;
+    private static final Injector injector = Injector.getInstance("mate");
+    private final DriverService driverService = (DriverService)
+            injector.getInstance(DriverService.class);
+    private final CarService carService = (CarService)
+            injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
```
antonEEmus, line 30: make overriden protected methods public

src/main/java/mate/controller/car/AllDriverFromCarController.java
```diff
@@ -0,0 +1,60 @@
...
+        Car car = carService.get(Long.parseLong(req.getParameter("id")));
+        req.setAttribute("car", car);
+        req.setAttribute("drivers", driverService.getAll());
+        req.getRequestDispatcher("/WEB-INF/views/cars/drivers.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        Map<String, String[]> parameterMap = req.getParameterMap();
+        Long carId = Long.parseLong(parameterMap.get("id")[INDEX_OF_FIRST_PARAM]);
```
antonEEmus, line 41: i think using map is a bad idea. better specify ids and get rid of map and constant indices

src/main/webapp/WEB-INF/views/cars/all.jsp
```diff
@@ -0,0 +1,39 @@
...
+            <td colspan="2">Setting</td>
+        </tr>
+        <c:forEach items="${cars}" var = "car">
+            <tr>
+                <td><c:out value="${car.id}"/></td>
+                <td><c:out value="${car.model}"/></td>
+                <td><c:out value="${car.manufacturer.name}"/></td>
+                <td><c:out value="${car.manufacturer.country}"/></td>
+                <td>
+                    <c:if test="${car.drivers.size() == 0}">free</c:if>
```
antonEEmus, line 26: nice feature, but "free" sounds like "free to buy/get". better write here "no drivers" or something.

src/main/webapp/WEB-INF/views/cars/all.jsp
```diff
@@ -0,0 +1,39 @@
...
+                    <c:if test="${car.drivers.size() != 0}">
+                        <c:forEach items="${car.drivers}" var = "driver">
+                            <c:out value="${driver.name}"/>: <c:out value="${driver.licenseNumber}"/><br>
+                        </c:forEach>
+                    </c:if>
+                </td>
+                <td><a href="${pageContext.request.contextPath}/cars/drivers?id=${car.id}">change drivers</a> </td>
```
antonEEmus, line 33: original)

src/main/webapp/WEB-INF/views/cars/drivers.jsp
```diff
@@ -0,0 +1,21 @@
...
+car model: ${car.model}<br>
+car manufacturer: ${car.manufacturer.name}<br>
+<form method="post" action="${pageContext.request.contextPath}/cars/drivers">
+    <c:forEach items="${drivers}" var = "driver">
+        <input type="checkbox" id="${driver.id}" name="${driver.id}" <c:if test="${car.drivers.contains(driver)}">checked</c:if>>
+        <label for="${driver.id}">${driver.name}: ${driver.licenseNumber}</label><br>
+    </c:forEach>
+    <input type="hidden" name="id" value="${car.id}" />
+    <button type="submit">Confirm</button>
+</form>
```
antonEEmus, line 19: what is this page for? diplaying drivers of car is not in this task) especially you don't need any form)

antonEEmus (changes requested):

src/main/webapp/WEB-INF/views/cars/drivers.jsp
zim10093: it is about the last TODO things in the readme file
"Create a new controller: Implement adding driver to car."
and when I see this picture in common mistake 
image
I try to create a page with the ability to change drivers in the car.
In my opinion, this page is more useful than 3 different pages where you can see all drivers, add only one or remove only one
Plus I think it improves performance because when we update a car we delete all drivers and then add all driver to DB

antonEEmus (approved):
good job!

src/main/java/mate/controller/car/AllDriversFromCarController.java
```diff
@@ -0,0 +1,62 @@
...
+    private final CarService carService = (CarService)
+            injector.getInstance(CarService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        if (req.getParameter("id") == null) {
+            resp.sendRedirect(req.getContextPath() + "/cars");
+            return;
+        }
```
dSkarzh, line 34: it's nice that you added this check, but let's keep logic simple and just remove it, it's redundant

dSkarzh (changes requested):
Well done! Very good solution!

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        Manufacturer manufacturer = manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturerId")));
+
+        car.setManufacturer(manufacturer);
+        car.setDrivers(Collections.emptyList());
```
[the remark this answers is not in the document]
zim10093: I know that it is immutable. 
in this case, we need to create a car with empty drivers and write it to DB.
when we fetch a car from DB it will make with the mutable list.
So for this case, i think I can use it

src/main/java/mate/controller/car/DeleteCarController.java
```diff
@@ -0,0 +1,23 @@
...
+import mate.service.CarService;
+
+@WebServlet(urlPatterns = "/cars/delete")
+public class DeleteCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
```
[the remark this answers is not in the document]
zim10093: Use inspect code to find reductant throws

dSkarzh (approved):
🔥

PR #270, opened 2022-07-11
Controllers were implemeted

mykola8426:


src/main/webapp/web.xml
```diff
@@ -11,4 +11,84 @@
...
+        <servlet-name>deleteManufacturer</servlet-name>
+        <url-pattern>/manufacturers/delete</url-pattern>
+    </servlet-mapping>
+    <servlet>
+        <servlet-name>deleteCar</servlet-name>
+        <servlet-class>mate.controller.DeleteCarController</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>deleteCar</servlet-name>
+        <url-pattern>/car/delete</url-pattern>
```
MKaramavrov, line 76:
```suggestion
        <url-pattern>/cars/delete</url-pattern>
```

MKaramavrov (changes requested):
Great!

MKaramavrov (approved):

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars/addDriverToCar.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
```
KostyaNikiforov, line 28: I guess you can remove ServletException, because you don't use the `forward` method here

src/main/java/mate/controller/GetAllManufacturersController.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final Injector injector = Injector.getInstance("mate");
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        req.setAttribute("manufacturers", manufacturers);
+        req.getRequestDispatcher("/WEB-INF/views/manufacturers/getAllManufacturers.jsp")
```
KostyaNikiforov, line 23: It is better to name jsp files simpler.
The path can look like this, for example: 
/manufacturers/all
/cars/delete
/cars/drivers/add

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
Good job:fire:

PR #277, opened 2022-07-12
Web-practice was finished

YaroslavLa:


DimaPodakin (approved):

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Manufacturer manufacturer =
+                manufacturerService.get(Long.valueOf(req.getParameter("manufacturer_id")));
+        String model = req.getParameter("model");
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
KostyaNikiforov, line 38: You also can use Collections.emptyList(), if you need to just initialize an empty collection and you are not going to add any elements

src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/drivers/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
```
KostyaNikiforov, line 27: I guess you can remove ServletException here, because you don't use a `forward` method in the method

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #278, opened 2022-07-13
created controllers

DimaPodakin:


src/main/java/mate/controller/CreateManufacturerController.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Manufacturer;
+import mate.service.ManufacturerService;
+
+public class CreateManufacturerController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private Manufacturer manufacturer;
```
YaroslavLa, line 16: You can move this variable to doPost method

YaroslavLa (approved):
good solution:metal:

src/main/java/mate/controller/CreateManufacturerController.java
KostyaNikiforov: It is not a part of the controller's dependencies. Controllers should be stateless, what if you get two requests from different users at the same time?

KostyaNikiforov (changes requested):

PR #281, opened 2022-07-13
Task completed

vitalii-smahlenko:


src/main/java/mate/Main.java
```diff
@@ -0,0 +1,6 @@
+package mate;
+
+public class Main {
```
MaliukDaria, line 3: Why do we need this class?

MaliukDaria (changes requested):

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,43 @@
...
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        Car car = new Car(model, manufacturer);
+        carService.create(car);
+        resp.sendRedirect("/index");
+    }
+
+    @Override
+    public void destroy() {
+        super.destroy();
+    }
```
MaliukDaria, line 42: why do we need this?

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturer_id"));
+        String model = req.getParameter("model");
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        Car car = new Car(model, manufacturer);
+        carService.create(car);
+        resp.sendRedirect("/index");
```
MaliukDaria, line 36: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.

src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String name = req.getParameter("name");
+        String licenseNumber = req.getParameter("license_number");
+        Driver driver = new Driver(name, licenseNumber);
+        driverService.create(driver);
+        resp.sendRedirect("/index");
```
MaliukDaria, line 32: If you use sendRedirect() method in your controllers, please pass "request.getContextPath()" + "/your-endpoint" as a parameter. Currently, the context path is empty, but if it is not, your code still should work.
And so on

MaliukDaria (changes requested):

PR #284, opened 2022-07-19
Solution added

Rommelua:


src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,36 @@
...
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/drivers/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String name = req.getParameter("name");
+        String licenseNumber = req.getParameter("license");
```
Nick97-git, line 29:
```suggestion
        String licenseNumber = req.getParameter("license_number");
```

src/main/java/mate/dao/CarDaoImpl.java
```diff
@@ -161,7 +161,7 @@ public List<Car> getAllByDriver(Long driverId) {
     private void insertAllDrivers(Car car) {
         Long carId = car.getId();
         List<Driver> drivers = car.getDrivers();
-        if (drivers.size() == 0) {
+        if (drivers == null || drivers.size() == 0) {
```
Nick97-git, line 164: return the original version, if I'm not wrong it's impossible that we will get null here

Nick97-git (changes requested):

src/main/java/mate/controller/AddDriverController.java
Rommelua: fixed

src/main/java/mate/dao/CarDaoImpl.java
Rommelua: fixed

Nick97-git (approved):

PR #286, opened 2022-07-20
Wp hw

alexgash8:


alexgash8:
I really wonder how to display cars with drivers in jsp, but can't figure it out by myself. Could you be so kind to give me the key idea?

PR #290, opened 2022-07-28
Hw jv web practice

Oleksii-Mospan:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,33 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        List<Driver> drivers = car.getDrivers();
+        drivers.add(driver);
+        car.setDrivers(drivers);
+        carService.update(car);
```
MaliukDaria, line 30: I think we have special method addDriverToCar in carService

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car car = new Car();
+        String model = req.getParameter("model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturerId"));
+        Manufacturer manufacturer = manufacturerService.get(manufacturerId);
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
MaliukDaria, line 40: we can set empty list to drivers in Car constructor

src/main/java/mate/controller/DisplayAllCarsController.java
```diff
@@ -0,0 +1,26 @@
...
+    private final CarService carsService =
+            (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.setAttribute("cars", carsService.getAll());
+        req.setAttribute("drivers", driverService.getAll());
```
MaliukDaria, line 23: Why do we need all drivers in DisplayAllCarsController?

MaliukDaria (changes requested):

src/main/java/mate/controller/DisplayAllCarsController.java
Oleksii-Mospan: As I am newbie in webapps and JSP, I was very curious as to how I can put a list of drivers next to the car. It was like a challenge for me :) And I think, that I done it. ))

Oleksii-Mospan: image

src/main/java/mate/model/Car.java
```diff
@@ -9,6 +9,13 @@ public class Car {
     private Manufacturer manufacturer;
     private List<Driver> drivers;
 
+    public Car() {
+    }
+
+    public Car(List<Driver> drivers) {
+        this.drivers = drivers;
+    }
```
MaliukDaria, line 17:
```suggestion
    public Car() {
      drivers = new ArrayList();
    }
```

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,43 @@
...
+            throws ServletException, IOException {
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        req.setAttribute("manufacturers", manufacturers);
+        req.getRequestDispatcher("/WEB-INF/views/cars/create.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car car = new Car(new ArrayList<>());
```
MaliukDaria, line 34:
```suggestion
        Car car = new Car();
```

src/main/java/mate/controller/DisplayAllCarsController.java
MaliukDaria, line 23: can you get to list of drivers through cars? As you get cars with drivers

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #314, opened 2022-09-11
added controllers

Litvik21:


amarkelovUA (approved):
It is surprising, but even in such a specific task, we have different markup

src/main/java/mate/controller/car/AddDriverIntoCarController.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars/drivers/add")
+public class AddDriverIntoCarController extends HttpServlet {
```
Nick97-git, line 16:
```suggestion
public class AddDriverToCarController extends HttpServlet {
```

src/main/java/mate/controller/car/CreateCarController.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add")
+public class CreateCarController extends HttpServlet {
```
Nick97-git, line 17:
```suggestion
public class AddCarController extends HttpServlet {
```
if you use verb `add` in url, it's better to use the same verb in controller's name, fix it everywhere

Nick97-git (changes requested):

Nick97-git (approved):

PR #365, opened 2022-09-23
Jv olesya khalak

olesyakhalakk:


src/main/java/mate/dao/CarDaoImpl.java
```diff
@@ -37,7 +37,7 @@ public Car create(Car car) {
                 car.setId(resultSet.getObject(1, Long.class));
             }
         } catch (SQLException e) {
-            throw new DataProcessingException("Can't create car " + car, e);
+            throw new DataProcessingException("Can't create mate.controller.car " + car, e);
```
aaananas, line 40:
```suggestion
            throw new DataProcessingException("Can't create car " + car, e);
```
why you added this everywhere

aaananas (approved):

PR #367, opened 2022-09-24
develop

voitalik:


src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,47 @@
...
+        req.getRequestDispatcher("/WEB-INF/views/cars/add.jsp")
+                .forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        String manufacturerId = req.getParameter("manufacturerId");
+        if (!model.isEmpty() && !manufacturerId.isEmpty()) {
```
kozhukhovsky, line 35: I think this check is redundant here.
And why is it only in some controllers?

kozhukhovsky (approved):

PR #388, opened 2022-11-11
implemented all tasks

nazariipertsak:


EvgenyTyshchenko (approved):
Mayby you should rename web views from all to getAllCars. because you use create car in cars folder

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService)
+            injector.getInstance(CarService.class);
+    private final DriverService driverService = (DriverService)
+            injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars/drivers/addDriverToCar.jsp")
```
KostyaNikiforov, line 26:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/cars/drivers/add.jsp")
```
we already see this part - `cars/drivers`

src/main/java/mate/controller/GetAllDriversController.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final Injector injector = Injector.getInstance("mate");
+    private final DriverService driverService = (DriverService)
+            injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Driver> drivers = driverService.getAll();
+        req.setAttribute("drivers", drivers);
+        req.getRequestDispatcher("/WEB-INF/views/drivers/getAllDrivers.jsp").forward(req, resp);
```
KostyaNikiforov, line 25:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/drivers/all.jsp").forward(req, resp);
```

src/main/java/mate/controller/GetAllManufacturersController.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final Injector injector = Injector.getInstance("mate");
+    private ManufacturerService manufacturerService = (ManufacturerService)
+            injector.getInstance(ManufacturerService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        req.setAttribute("manufacturers", manufacturers);
+        req.getRequestDispatcher("/WEB-INF/views/manufacturers/getAllManufacturers.jsp")
```
KostyaNikiforov, line 25:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/manufacturers/all.jsp")
```

src/main/java/mate/model/Car.java
```diff
@@ -1,13 +1,14 @@
...
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
 public class Car {
     private Long id;
     private String model;
     private Manufacturer manufacturer;
-    private List<Driver> drivers;
+    private List<Driver> drivers = new ArrayList<>();
```
KostyaNikiforov, line 11: An entity can be with not initialized files. There can be a confusing situation, or we get a car without drivers or the drivers field is just not initialized. Let's leave `null` in case it was not initialized

KostyaNikiforov: Also, there is a rule that initialization logic must be inside the constructor or setter method

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #396, opened 2022-11-16
Home work

roman680:


src/main/java/mate/controller/IndexController.java
```diff
@@ -1,15 +1,18 @@
...
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import mate.util.ConnectionUtil;
 
 public class IndexController extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
+        Connection connection = ConnectionUtil.getConnection();
```
aaananas, line 15: why do you need it here

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,58 @@
...
+        car.setManufacturer(getManufacturer(manufacturer));
+        car.setDrivers(List.of());
+        return car;
+    }
+
+    private Manufacturer getManufacturer(String manufacturer) {
+        return manufacturerService.getAll().stream()
+                 .filter(manufactur -> manufactur.getName().equalsIgnoreCase(manufacturer))
+                 .findAny().orElseThrow(()
+                         -> new RuntimeException("Cant wind this manufacturer in DB"));
```
aaananas, line 56: in form you can pass id, and get manufacturer by id, instead of filtering the whole list

aaananas (changes requested):

aaananas (approved):

PR #405, opened 2022-11-21
Impl Controller's level (Car, Driver, Manufacturer). Also add several…

IvanAldokhin:
… jsp ( get, add, delete).

### DIFF
--- a/src/main/java/mate/controller/IndexController.java
+++ b/src/main/java/mate/controller/IndexController.java
@@ -10,6 +10,7 @@ public class IndexController extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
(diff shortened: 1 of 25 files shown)


src/main/java/mate/controller/IndexController.java
```diff
@@ -10,6 +10,7 @@ public class IndexController extends HttpServlet {
     @Override
     public void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
-        req.getRequestDispatcher("/WEB-INF/views/index.jsp").forward(req, resp);
+        req.getRequestDispatcher(req.getContextPath()
+                + "/WEB-INF/views/index.jsp").forward(req, resp);
```
aaananas, line 14: contextPath should be added only in sendRedirect() method

aaananas (changes requested):

aaananas (changes requested):

kozhukhovsky (approved):

PR #431, opened 2022-11-26
solved jv web practice task

HryhorashPavlo:


src/main/webapp/WEB-INF/views/car/create.jsp
```diff
@@ -0,0 +1,17 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>Create car</title>
+</head>
+<body>
+<h1>Create car</h1>
+<h4 style="color: green">${message}</h4>
+<form method="post" action="${pageContext.request.contextPath}/cars/create">
```
Veronika-Tretiak, line 9: /cars/add

src/main/webapp/WEB-INF/views/driver/add.jsp
```diff
@@ -0,0 +1,17 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>Add driver to the car</title>
+</head>
+<body>
+<h1>Add driver to the car</h1>
+<h4 style="color: green">${message}</h4>
+<form method="post" action="${pageContext.request.contextPath}/cars/drivers/add">
```
Veronika-Tretiak, line 9: I think it is better to transfer this file to the car package and and create there one more package driver and  add this file.
car/driver/add
Check all naming create -add
display-all - all

Veronika-Tretiak (approved):

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,36 @@
...
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher(PATH).forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        carService.addDriverToCar(driverService.get(Long.parseLong(req.getParameter("driver_id"))),
+                carService.get(Long.parseLong(req.getParameter("car_id"))));
```
ChumakAn, line 32: It's better to separate this to avoid heavy-looking and difficult to read code

ChumakAn (approved):
Good job, but take into account my comment for the next HW😉

src/main/java/mate/controller/AddDriverToCarController.java
HryhorashPavlo: Ok, got it

PR #443, opened 2023-01-08
solution to jv-web-practice

romanovosad87:


src/main/java/mate/controller/DeleteCarController.java
```diff
@@ -0,0 +1,23 @@
...
+import mate.lib.Injector;
+import mate.service.CarService;
+
+@WebServlet(urlPatterns = "/cars/delete")
+public class DeleteCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
```
nacenik, line 18: Maybe you need to send a request after successfully deleting it? As a variant - send request to see all cars. Fix it in all places.

pom.xml
```diff
@@ -23,14 +23,19 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.22</version>
+            <version>8.0.30</version>
```
nacenik, line 26: don't change connector version

nacenik (changes requested):

aleksandr-bevza (approved):

PR #444, opened 2023-01-15
Added Controllers and JSP files to project.Also added web.xml and som…

AntonKalenskiy:
…e dependencies into pom.xml.


Ivan95kos (approved):
Great work

PR #445, opened 2023-01-15
Hw jv web practice

DanyloBorysenko:


Ivan95kos:
fix checkstyle

Ivan95kos (changes requested):
fix checkstyle

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,34 @@
...
+        req.getRequestDispatcher("/WEB-INF/views/cars/add_driver.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long carId = Long.valueOf(req.getParameter("carId"));
+        Long driverId = Long.valueOf(req.getParameter("driverId"));
+        carService.addDriverToCar(driverService.get(driverId), carService.get(carId));
+        resp.sendRedirect("/confirmation");
```
aleksandr-bevza, line 32: Fix it in all places:
image
```suggestion
        resp.sendRedirect(req.getContextPath() + "/confirmation");
```

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,33 @@
...
+import mate.service.CarService;
+
+@WebServlet(urlPatterns = "/cars/add")
+public class CreateCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
```
aleksandr-bevza, line 20: It would be better to give user possibility to select manufacturer from the list.
```suggestion
            throws ServletException, IOException {
        List<Manufacturer> allManufacturers = manufacturerService.getAll();
        req.setAttribute("manufacturers", allManufacturers);
```

src/main/webapp/WEB-INF/views/cars/car_registration.jsp
```diff
@@ -0,0 +1,14 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>Add new car</title>
+</head>
+<body>
+<h1>Car registration form</h1>
+<form method="post" action="${pageContext.request.contextPath}/cars/add">
+    MODEL <input type="text" name="model"><br>
+    MANUFACTURER ID <input type="text" name="manufacturer_id"><br>
```
aleksandr-bevza, line 10: It would be better to give user possibility to select manufacturer from the list.
```suggestion
    MANUFACTURER ID: <br/>
    <c:forEach items="${manufacturers}" var="manufacturer">
        <input type="radio" id="html" name="manufacturer_id" value="${manufacturer.id}" required>
        <label for="${manufacturer.id}">${manufacturer.name}</label><br>
    </c:forEach>
    <br/>
```

src/main/java/mate/service/CarService.java
```diff
@@ -10,4 +10,6 @@ public interface CarService extends GenericService<Car> {
     void removeDriverFromCar(Driver driver, Car car);
 
     List<Car> getAllByDriver(Long driverId);
+
+    Car registerNewCar(String model, Long manufacturerId);
```
aleksandr-bevza, line 14: This logic is confusing. You already have method 'create' that saves car to the database. And you have method 'register' that creates new car but does not save it into database. It's not obvious from method's name, what does it do. 
In this homework it's better not to create 'register' method. Create new Car in Controller.
Also remove 'register' method from other Services

src/main/webapp/WEB-INF/views/info/confirmation.jsp
```diff
@@ -1,9 +1,9 @@
-<%@ page contentType="text/html;charset=UTF-8" language="java" %>
-<html>
-<head>
-    <title>My team</title>
-</head>
-<body>
-<h1>Hello, mates!</h1>
-</body>
-</html>
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
```
aleksandr-bevza, line 1: It's better to have main page in your application: index.jsp

aleksandr-bevza (changes requested):

src/main/java/mate/service/CarService.java
DanyloBorysenko: I thought it is a good idea, if we will want to add some check style for inputed data.

Ivan95kos (approved):
Great work

PR #455, opened 2023-01-31
my web practice

InnaDiachkina:


src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.DriverService;
+
+public class AddDriverController extends HttpServlet {
```
MaxMaksym, line 12: It's ok to use mapping, but I think that annotation @WebServlet is better for readability. Just a tip, you can use it in the next hw)

src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Driver driver = new Driver();
+        driver.setName(req.getParameter("name"));
+        driver.setLicenseNumber(req.getParameter("license_number"));
+        driverService.create(driver);
+        resp.sendRedirect(req.getContextPath() + "/drivers/add");
```
MaxMaksym, line 30: same

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        Long manufacturerId = Long.parseLong((req.getParameter("manufacturer_id")));
+        car.setManufacturer(manufacturerService.get(manufacturerId));
+        car.setDrivers(new ArrayList<>());
+        carService.create(car);
+        resp.sendRedirect(req.getContextPath() + "/cars/add");
```
MaxMaksym, line 35: think about adding some message to jsp page which informs user that changes was saved, or you can redirect user to page with this message

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,37 @@
...
+        req.getRequestDispatcher("/WEB-INF/views/cars/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        Long manufacturerId = Long.parseLong((req.getParameter("manufacturer_id")));
+        car.setManufacturer(manufacturerService.get(manufacturerId));
```
MaxMaksym, line 32: what if you don't have manufacturer with this id? you can add one more method in  ManufacurerService which returns Optional of manufacturer and in case it's empty- add message to jsp page that manufacurer does not exist

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,37 @@
...
+        req.getRequestDispatcher("/WEB-INF/views/cars/addDriver.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
```
MaxMaksym, line 32: Same if you don't have car of driver in db

MaxMaksym (changes requested):
image
What about this servlets? Lets ask Mykyta tomorrow if they are optional or you still need to implement them)

src/main/java/mate/controller/AddDriverToCarController.java
InnaDiachkina: Remember that you can pass id of a manufacturer, car, or driver through the form where your models depend on one another. You can expect that the entity with this id already exists in the DB.

src/main/java/mate/controller/AddCarController.java
InnaDiachkina: Remember that you can pass id of a manufacturer, car, or driver through the form where your models depend on one another. You can expect that the entity with this id already exists in the DB.

MaxMaksym (approved):

Nick97-git (approved):

PR #468, opened 2023-02-02
all necessary tasks were implemented

lysytskapolina:


Pandoryms (approved):

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        car.getDrivers().add(driver);
+        carService.update(car);
```
Nick97-git, line 34: you have addDriverToCar() method, use it

Nick97-git (changes requested):

src/main/java/mate/controller/AddDriverToCarController.java
lysytskapolina: You're right, it's much easier😅

Nick97-git (approved):

PR #479, opened 2023-02-05
Hw

Mykola-Osolinskyi:


Serhii1206 (approved):

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/drivers/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
```
KostyaNikiforov, line 32: remove unused exceptions from the method signature
fix it everywhere

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/create")
```
KostyaNikiforov, line 16:
```suggestion
@WebServlet("/cars/create")
```

KostyaNikiforov (changes requested):

src/main/java/mate/controller/AddDriverToCarController.java
Mykola-Osolinskyi: we use method forward. Looks like, we need both exceptions
when I delete one of them, I get compile time error

KostyaNikiforov (approved):

Serhii1206 (approved):

PR #495, opened 2023-02-22
feat: jv-web-practice #comment was added webapp to taxi service

sania517:


src/main/java/mate/controller/cars/AddDriverToCarController.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new RuntimeException("Can not delete driver from undefined Car. carId = "
+                    + carId);
+        }
+
+        if (driverId == null || driverId.length() < 1) {
+            throw new RuntimeException("Can not delete driver with empty driverId. driverId = "
+                    + driverId);
+        }
+
+        carService.addDriverToCar(
```
olekskov, line 37: Please make assignment process to be performed in POST method

olekskov (changes requested):
Good job! Let's polish some points highlighted in comments

src/main/java/mate/controller/cars/AddDriverToCarController.java
sania517: For using the POST method I must create a form. In my implementation, I add a driver by link. If it is mandatory I can create a crutch  - put form without fields instead of the link.

Ivan95kos: So it is necessary because it violates the principles of REST, 
GET method is mainly for getting data and POST is for creating
https://dou.ua/forums/topic/34550/

Ivan95kos (changes requested):

Ivan95kos (approved):
Great job

PR #499, opened 2023-03-01
Here is my solution for hw-web-practice. Please start with "/index" page

andmalenko:


Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/mate/controller/GetAllCarsController.java
```diff
@@ -0,0 +1,24 @@
...
+public class GetAllCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        CarService carService =
+                (CarService) injector.getInstance(CarService.class);
+        req.setAttribute("cars", carService.getAll());
+        req.getRequestDispatcher("/WEB-INF/views/car/all.jsp").forward(req, resp);
```
aaananas, line 22:
```suggestion
    private static final Injector injector = Injector.getInstance("mate");
    private final CarService carService =
                (CarService) injector.getInstance(CarService.class);

    @Override
    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
            throws ServletException, IOException {
        req.setAttribute("cars", carService.getAll());
        req.getRequestDispatcher("/WEB-INF/views/car/all.jsp").forward(req, resp);
```
initialize as a class-level field - fix all controllers

aaananas (changes requested):

aaananas (approved):

PR #501, opened 2023-03-04
1

dimonkoval:


src/main/java/mate/controller/DeleteCarController.java
```diff
@@ -0,0 +1,34 @@
...
+        response.sendRedirect(request.getContextPath() + "/cars/delete");
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Car> currentCars = carService.getAll();
+        req.setAttribute("cars", currentCars);
+        req.getRequestDispatcher("/WEB-INF/views/cars/delete.jsp").forward(req, resp);
+    }
```
olekskov, line 33: In **all** delete operations/controllers I don't like an idea that they are combined with get controllers.
Please split this logic to corresponding controllers.
GetAll...Controller will have doGet method and provide user list of some objects, Delete...Controller will have doPostMethod, perform deletion logic and will redirect user to respective GetAll page

Because now it's really strange that after object creation we redirect user to delete page where he gets list of objects

olekskov (changes requested):
Please check comment below

kozhukhovsky (approved):

PR #505, opened 2023-03-10
Implement JSP pages and their controllers

katerynapoli:


okuzan (changes requested):
Create packages - car, driver, manufacturer in controller package

okuzan (approved):
Well done!

PR #507, opened 2023-03-16
Solution

kateryna-mykh:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp) 
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
+        resp.sendRedirect(req.getContextPath() + "/index");
```
olekskov, line 37: It's strange that we are redirecting user to index page every time, let's send user to page where he can see result of his action (related getAll page). Pls recheck all controllers

pom.xml
olekskov, line 1: It doesn't look like necessary to change margins for the whole file

kateryna-mykh: I fixed it. (It is already a habit to apply formatting to the entire file.)

okuzan (approved):
Well done!
could also divide controllers into packages

PR #513, opened 2023-03-19
1st try

Watxesnavo:


src/main/java/mate/controller/DeleteDriverController.java
```diff
@@ -0,0 +1,25 @@
...
+import mate.service.DriverService;
+
+public class DeleteDriverController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final DriverDao driverDao = (DriverDao) injector.getInstance(DriverDao.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
```
olekskov, line 19: Please rework method to work via POST method, get should be used for obtaining information, not modification

src/main/java/mate/controller/DeleteManufacturerController.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class DeleteManufacturerController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final ManufacturerDao dao =
+            (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
```
olekskov, line 20: Please rework method to work via POST method, get should be used for obtaining information, not modification

src/main/java/mate/controller/DisplayAllCarsController.java
```diff
@@ -0,0 +1,37 @@
...
+        List<Car> cars = carService.getAll();
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        List<Driver> drivers = null;
+        for (Car car : cars) {
+            manufacturers.add(car.getManufacturer());
+            drivers = car.getDrivers();
+        }
+        req.setAttribute("drivers", drivers);
+        req.setAttribute("manufacturers", manufacturers);
+        req.setAttribute("cars", cars);
```
olekskov, line 34: Sorry, but I really don't understand what's going on here.
We have full cars data obtained from carService.getAll() method but instead of passing it as is to jsp page to be properly displayed we starting to extract included properties to additional list, same time drivers are overridden in cycle instead of being added...
I offer to rework jsp page to work with just list of cars and get all required info from there

olekskov (changes requested):

kozhukhovsky (approved):

PR #528, opened 2023-04-04
Hw web practice

Iskamele:
I took the liberty of digressing a bit from a task and adding css styles for implementation.
And additionally, I added dropdowns where they can be added to make it easier to use the implementation. This is probably not a very good solution, but it was helpful for practice 😁
I also had some difficulties with the URL naming, so if there are any comments, I will be glad to hear them.

### DIFF
--- a/src/main/java/mate/controller/car/CreateCarController.java
+++ b/src/main/java/mate/controller/car/CreateCarController.java
@@ -1,0 +1,53 @@
+package mate.controller.car;
+
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.DriverService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add")
+public class CreateCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+    private final ManufacturerService manufacturerService =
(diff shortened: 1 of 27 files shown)


src/main/java/mate/controller/car/CreateCarController.java
```diff
@@ -0,0 +1,53 @@
...
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.model.Manufacturer;
+import mate.service.CarService;
+import mate.service.DriverService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add")
+public class CreateCarController extends HttpServlet {
```
Rommelua, line 19: Better use the same word for the same action everywhere. **Add** everywhere or **create** everywhere.
**For all controllers.**
```suggestion
@WebServlet(urlPatterns = "/cars/add")
public class AddCarController extends HttpServlet {
```

src/main/java/mate/controller/driver/CreateDriverController.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/drivers/add")
+public class CreateDriverController extends HttpServlet {
```
Rommelua, line 14: add

src/main/java/mate/controller/manufacturer/CreateManufacturerController.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Manufacturer;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/manufacturers/add")
+public class CreateManufacturerController extends HttpServlet {
```
Rommelua, line 14: add

Rommelua (approved):

PR #529, opened 2023-04-04
implemented all controllers and jsps

OlehSarapuk:


src/main/java/mate/controller/IndexController.java
```diff
@@ -11,5 +11,6 @@ public class IndexController extends HttpServlet {
     public void doGet(HttpServletRequest req, HttpServletResponse resp)
             throws ServletException, IOException {
         req.getRequestDispatcher("/WEB-INF/views/index.jsp").forward(req, resp);
+        ;
```
IlkoGood, line 14: what is it? :)

src/main/webapp/web.xml
```diff
@@ -11,4 +11,94 @@
         <servlet-name>index</servlet-name>
         <url-pattern>/index</url-pattern>
     </servlet-mapping>
+
```
IlkoGood, line 14: Not sure this empty lines are needed

src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Driver driver = new Driver();
+        driver.setName(req.getParameter("name"));
+        driver.setLicenseNumber(req.getParameter("license_number"));
+        driverService.create(driver);
+        resp.sendRedirect(req.getContextPath() + "/drivers/all");
```
IlkoGood, line 29: When adding a car you return to index, when creating driver - to the list of drivers, seems strange.

IlkoGood (approved):
Working model, approved.
It would be nice to have some navigation on the index page.
Also Home button on all the pages would be nice.

src/main/java/mate/controller/CreateDriverController.java
petro-ushchuk: agree, u should chose one style

src/main/java/mate/controller/IndexController.java
petro-ushchuk: fix this ;)

petro-ushchuk (changes requested):

nacenik (approved):
LGTM! 👍

PR #538, opened 2023-04-05
added controllers

IvanZuravlev228:


V-Vasylchuk (approved):

src/main/webapp/WEB-INF/views/index.jsp
```diff
@@ -4,6 +4,6 @@
     <title>My team</title>
 </head>
 <body>
-<h1>Hello, mates!</h1>
+    <h1>Hello, mates!</h1>
```
nacenik, line 7: As improvements - it's a good idea to create a menu from the index page to other pages, cause users might not know which menu your application has.

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #548, opened 2023-04-05
created solution 

ArtemGrunin:


RandomEastEcho (approved):

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/cars/add-driver")
```
petro-ushchuk, line 15: Check common mistakes file and fix your url naming

petro-ushchuk (changes requested):
Good work, let's fix one common mistake and we are done 🔥

src/main/java/mate/controller/AddDriverToCarController.java
ArtemGrunin: fixed)) thank you))

petro-ushchuk (approved):
Good work 🔥

PR #553, opened 2023-04-06
Web Practice

Deschna:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(request.getParameter("driver_id"));
+        Long carId = Long.valueOf(request.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
+    }
```
KosMarch, line 37: It would be better to add redirect, but it is not necessary
image

KosMarch (approved):
Good job!

karma-o (approved):
LGTM 👍

PR #554, opened 2023-04-06
Hw-practice

KosMarch:


src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car newCar = new Car();
+        newCar.setModel(req.getParameter("model"));
+        newCar.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturer_id"))));
+        newCar.setDrivers(Collections.emptyList());
```
Deschna, line 35: Collections.emptyList() return immutable list, what means you wont be able to add new drivers to it

Deschna (changes requested):
Overall good except the one issue

Deschna (approved):

src/main/java/mate/controller/AddCarController.java
karma-o: That's fine since we only need to create a car without its drivers here

KosMarch: I checked the code, everything seems to work in practice

karma-o (approved):
looks good

PR #564, opened 2023-04-08
new controllers implemented

AnonimToster:


VitaliiPotseluiko (approved):

src/main/webapp/WEB-INF/views/cars-drivers.jsp
```diff
@@ -0,0 +1,41 @@
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
+<html>
+<head>
+    <title>Car drivers</title>
+</head>
+    <body>
+        <%@ include file="/WEB-INF/views/header.jsp"%><br>
+        <c:if test="${drivers.size() == 0}">No cars exist!</c:if>
```
nacenik, line 9:
```suggestion
        <c:if test="${drivers.size() == 0}">No drivers exist!</c:if>
```

src/main/java/mate/controller/car/driver/GetCarDriversController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String id = req.getParameter("id");
+        if (id != null) {
+            Car car = carService.get(Long.valueOf(id));
+            List<Driver> drivers = car.getDrivers();
+            List<Driver> candidates = driverService.getAll();
+            candidates.removeAll(drivers);
```
nacenik, line 32: this is business logic. Better to put it into service layer

nacenik (approved):
All tasks are done, good job! 👍 

Note: Be careful with business logic inside controller

PR #574, opened 2023-04-28
Jv web practice solution

mrGvvD:


src/main/java/mate/controller/AllDriversController.java
```diff
@@ -0,0 +1,24 @@
...
+public class AllDriversController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private static final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.setAttribute("drivers", driverService.getAll());
+        req.getRequestDispatcher("/WEB-INF/views/drivers/allDrivers.jsp").forward(req,resp);
```
Ivan95kos, line 22: Add a space after the comma in the 'forward()' method.

src/main/java/mate/controller/DeleteManufacturerController.java
```diff
@@ -0,0 +1,28 @@
...
+public class DeleteManufacturerController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private static final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long id = Long.parseLong(req.getParameter("id"));
+        if (manufacturerService.delete(id)) {
```
Ivan95kos, line 22: It is better to handle the case when the deletion fails. You can add an else block and return an error message.

src/main/java/mate/controller/DeleteDriverController.java
```diff
@@ -0,0 +1,28 @@
...
+public class DeleteDriverController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private static final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long id = Long.parseLong(req.getParameter("id"));
+        if (driverService.delete(id)) {
```
Ivan95kos, line 22: It's better to separate the action of deleting a driver and redirecting to the message box into different lines for better readability.

src/main/java/mate/controller/DeleteCarController.java
```diff
@@ -0,0 +1,27 @@
...
+@WebServlet(urlPatterns = "/cars/delete")
+public class DeleteCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector.getInstance(CarService.class);
+
+    @Override
+    public void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long id = Long.parseLong(req.getParameter("id"));
+        if (carService.delete(id)) {
```
Ivan95kos, line 21: It's better to separate the deletion and redirection into two separate lines for better readability.

src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,37 @@
...
+        req.getRequestDispatcher("/WEB-INF/views/drivers/createDriver.jsp").forward(req, resp);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Driver driver = new Driver();
+        driver.setName(req.getParameter("name"));
+        driver.setLicenseNumber(req.getParameter("licenseNumber"));
+        if (driverService.create(driver).getId() != null) {
```
Ivan95kos, line 31: Instead of checking if the driver's ID is not null after creating the driver, it is better to check if the driver object itself is not null. This will ensure proper error handling in case the 'create' method returns a null value.

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,41 @@
...
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars/addDriver.jsp").forward(req, resp);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Driver driver = driverService.get(Long.parseLong(req.getParameter("driverId")));
+        Car car = carService.get(Long.parseLong(req.getParameter("carId")));
+        carService.addDriverToCar(driver,car);
```
Ivan95kos, line 33: Insert a space after the comma when passing the arguments to the addDriverToCar() method.

src/main/java/mate/controller/CreateManufacturerController.java
```diff
@@ -0,0 +1,38 @@
...
+                .forward(req,resp);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setCountry(req.getParameter("country"));
+        manufacturer.setName(req.getParameter("name"));
+        if (manufacturerService.create(manufacturer).getId() != null) {
```
Ivan95kos, line 32: It's better to first store the result of the `create` method in a variable and then use the stored result in the conditional statement.

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        car.setManufacturer(manufacturerService.get(
+                Long.parseLong(req.getParameter("manufacturerId"))));
+        car.setDrivers(Collections.emptyList());
+        if (carService.create(car).getId() != null) {
```
Ivan95kos, line 36: It's better to check for success based on the returned Car object instead of checking its id. If a new Car object is returned, it means the operation was successful.

Ivan95kos (changes requested):
Good job overall!

kozhukhovsky (approved):

PR #577, opened 2023-05-12
My realisation of web taxi service

antonsablev:


src/main/java/mate/controller/DeleteController.java
```diff
@@ -0,0 +1,52 @@
...
+                break;
+
+            case DRIVER_CLASS:
+                driverService.delete(id);
+                break;
+
+            default:
+                break;
+
+        }
```
Ivan95kos, line 49: For the future, it is better to create three controllers than to use a switch

Ivan95kos (approved):
Great job, with minor comments

PR #579, opened 2023-05-18
added proper solution for this task

polikarpov13:


src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet("/driverToCar")
```
kozhukhovsky, line 15:
```suggestion
@WebServlet("/addDriverToCar")

```

It's not very clear what exactly.
It's not yet rest to give names to endpoints by essence. Therefore, since we have a separate servlet for each operation, it should be clear what kind of operation it is.

kozhukhovsky (approved):

PR #602, opened 2023-06-18
jv-web-practice/home-work

andrii-subash:


okuzan (changes requested):
Delete main, separtate controllers into packages

okuzan (changes requested):
image

olekskov (approved):
Good job

PR #606, opened 2023-06-19
My branch

V-Solianyk:


src/main/webapp/web.xml
```diff
@@ -3,6 +3,86 @@
...
+        <servlet-name>getAllDriversController</servlet-name>
+        <url-pattern>/drivers</url-pattern>
+    </servlet-mapping>
+    <servlet>
+        <servlet-name>createDriverController</servlet-name>
+        <servlet-class>mate.controller.CreateDriverController</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>createDriverController</servlet-name>
+        <url-pattern>/drivers/add</url-pattern>
```
Tupadzaky, line 20: again, problem with consistency. Your jsp files are called create.jsp, so they don't match to the url

src/main/webapp/web.xml
```diff
@@ -3,6 +3,86 @@
          xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
                       http://xmlns.jcp.org/xml/ns/javaee/web-app_4_0.xsd"
          version="4.0">
+    <servlet>
+        <servlet-name>getAllDriversController</servlet-name>
+        <servlet-class>mate.controller.GetAllDriversController</servlet-class>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>getAllDriversController</servlet-name>
+        <url-pattern>/drivers</url-pattern>
```
Tupadzaky, line 12: again, problem with consistency. url is correct, but your packages in views don't match, they should be named driver-s-, manufacturer-s- and car-s-.

Tupadzaky (changes requested):
The solution is good, but there is a couple of minor problems with naming, check them out in the comments

Tupadzaky (approved):
Good job!

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+public class AddDriverToCarController extends HttpServlet {
+    private static final String PACKAGE = "mate";
```
sarakhmen, line 15: there is no need to make this package name a constant, let's use it directly as a literal where needed. Please fix the same problem in all places

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final String PATH_TO_JSP = "/WEB-INF/views/cars/drivers/addDriversToCar.jsp";
```
sarakhmen:
```suggestion
    private static final String PATH_TO_JSP = "/WEB-INF/views/cars/drivers/add.jsp";
```
in order to make your jsp names consistent with urls

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        car.getDrivers().add(driver);
+        carService.update(car);
+        resp.sendRedirect(req.getContextPath() + "/index");
```
sarakhmen, line 37: is /index page the most appropriate one here?

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/mate/controller/AddDriverToCarController.java
V-Solianyk: "HINT: while implementing the last requirement, create just two fields on JSP where you will pass corresponding id values. You can redirect to /index page or any other page after your business logic in doPost() method is applied."
I took it from the task hint, but maybe I got it wrong?

PR #615, opened 2023-06-24
implemented task

chrisnyrka:


src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturer_id"))));
+        car.setModel(req.getParameter("model"));
+        car.setDrivers(Collections.emptyList());
```
m-pnsk, line 37: I think it would be better to put the setter parameters in separate variables, but this is just a recommendation

src/main/java/mate/dao/impl/DriverDaoImpl.java
```diff
@@ -38,7 +39,9 @@ public Driver create(Driver driver) {
 
     @Override
     public Optional<Driver> get(Long id) {
-        String query = "SELECT * FROM drivers WHERE id = ? AND is_deleted = FALSE";
+        String query = "SELECT id, name, license_number "
+                + "FROM drivers "
+                + "WHERE id = ? AND is_deleted = FALSE;";
```
m-pnsk, line 44: You shouldn't have changed the dao part

src/main/java/mate/model/Car.java
```diff
@@ -9,6 +9,10 @@ public class Car {
     private Manufacturer manufacturer;
     private List<Driver> drivers;
 
+    public Car() {
+
+    }
+
```
m-pnsk, line 15: This is redundant

m-pnsk (approved):
Well done. See my minor comments.

src/main/java/mate/controller/car/AddCarController.java
chrisnyrka: I'm not sure that it would be better. Let's see what the mentors have to say

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,41 @@
...
+import mate.model.Car;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet(urlPatterns = "/cars/add")
+public class AddCarController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService = (CarService) injector
+            .getInstance(CarService.class);
+
```
rovein, line 20: This empty line can be omitted.

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturer_id"))));
+        car.setModel(req.getParameter("model"));
+        car.setDrivers(new ArrayList<>());
```
rovein, line 37: To get rid of explicitly setting an empty list you can init `drivers` field in Car class.
`private List<Drivers> drivers = new ArrayList<>();`

Or you can create a constructor with `manufacturer` and `model` and init empty list here.

Feel free to choose any option.

src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,41 @@
...
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars/add.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car();
+        car.setManufacturer(manufacturerService
+                .get(Long.parseLong(req.getParameter("manufacturer_id"))));
```
rovein, line 35: Also, don't pass complex expressions to the constructor or setter instead try this:
```suggestion
        Manufacturer manufacturer = manufacturerService
                .get(Long.parseLong(req.getParameter("manufacturer_id")))
        Car car = new Car();
        car.setManufacturer(manufacturer);
```

src/main/java/mate/controller/car/AddDriverToCarController.java
```diff
@@ -0,0 +1,39 @@
...
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
+        resp.sendRedirect(req.getContextPath() + "/cars");
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
```
rovein, line 35: Let's change the ordering of the methods in this class. Firstly put `doGet`, then `doPost`. Verify in all places.

src/main/webapp/WEB-INF/views/cars/add.jsp
```diff
@@ -0,0 +1,15 @@
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
```
rovein, line 1: Can be removed because you don't use it in this file.
Check all cases.

rovein (changes requested):
Good job! But let's improve your solution ;)

sarakhmen (approved):
Good job ;)

PR #619, opened 2023-06-26
Homework: JV Web Practice

ViktorBykh:


maksigors (approved):

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,47 @@
+package mate.controller;
```
maxline, line 1: Let's create a separate packages `driver`, `car`, `manufacturer` inside `controller` package and divide controllers between these packages

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,47 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car createdCar = new Car();
+        String carModel = req.getParameter("car_model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturer_id"));
+        List<Driver> defaultDrivers = new ArrayList<>();
+        Manufacturer carManufacturer = manufacturerService.get(manufacturerId);
+        createdCar.setModel(carModel);
+        createdCar.setManufacturer(carManufacturer);
+        createdCar.setDrivers(defaultDrivers);
```
maxline, line 43: Try to not set an empty values when creating a new object. It is better initialize empty list in appropriate constructor in the `Car` class.
Also try to use constructor with parameters instead of default constructor with setter, this will simplify the code.

src/main/java/mate/controller/AddDriverController.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Driver createdDriver = new Driver();
+        String driverName = req.getParameter("driver_name");
+        String licenseNumber = req.getParameter("license_number");
+        createdDriver.setName(driverName);
+        createdDriver.setLicenseNumber(licenseNumber);
+        driverService.create(createdDriver);
```
maxline, line 34: Try to use constructor with parameters instead of default constructor with setters, this will simplify the code.

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,41 @@
...
+    private static final Injector injector = Injector.getInstance("mate");
+    private final CarService carService =
+            (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/cars_drivers/add_driver_to_car.jsp")
```
maxline, line 26:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/cars/add_driver_to_car.jsp")
```

src/main/java/mate/controller/DoneActionController.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.io.IOException;
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet(urlPatterns = "/done")
+public class DoneActionController extends HttpServlet {
```
maxline, line 11: I think this class is redundant. Let's better redirect to the previous page when we do some actions.
E.g. after creating cars will redirect to all cars page, after creating new driver - redirects to the all drivers page etc.

src/main/java/mate/controller/AddCarController.java
```diff
@@ -0,0 +1,47 @@
...
+        carService.create(createdCar);
+        resp.sendRedirect("/done");
```
maxline, line 45: let's better redirect to the all cars page
```suggestion
        resp.sendRedirect(req.getContextPath() + "/cars");
```
fix redirection please in the similar way in all controllers

maxline (changes requested):
Good job! Let's improve the solution

sarakhmen (approved):
Good job ;)

PR #627, opened 2023-06-26
 jv-web-practice

eugenetolkunov:


Valdemar619 (approved):
Well Done

src/main/java/mate/controller/car/CreateCarController.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws IOException {
+        Car car = new Car();
+        Manufacturer manufacturer = manufacturerService.get(
+                Long.valueOf(req.getParameter("manufacturer_id")));
+        car.setModel(req.getParameter("model"));
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
```
sarakhmen, line 37: To get rid of explicitly setting an empty list you can init add an empty list within the constructor or directly in the field.

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #633, opened 2023-06-27
added solution

Fiasko3304:


Pkrasota (approved):
LGTM!

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,41 @@
...
+            = Injector.getInstance("mate");
+    private final CarService carService
+            = (CarService) injector.getInstance(CarService.class);
+    private final DriverService driverService
+            = (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/views/car/driver/addDriver.jsp")
```
sarakhmen, line 27:
```suggestion
        req.getRequestDispatcher("/WEB-INF/views/cars/drivers/add.jsp")
```
I'd recommend following the same structure as for urls. Let's fix the same issues in all places

sarakhmen (changes requested):
Let's slightly improve your solution :)

rovein (approved):
Well done!

PR #635, opened 2023-06-27
solution

volodymyrruso:


src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,33 @@
...
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/drivers/create.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Driver driver = new Driver(req.getParameter("name"), req.getParameter("licenseNumber"));
+        resp.sendRedirect(req.getContextPath() + "/drivers");
```
DV7Art, line 29: I think it's unnecessary.

src/main/java/mate/controller/CreateDriverController.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Driver;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/drivers/create")
```
DV7Art, line 13: to create a car u use "/cars/add", to create a driver u use "/drivers/create" ~?

src/main/webapp/WEB-INF/drivers/drivers.jsp
```diff
@@ -0,0 +1,23 @@
+<%@ page contentType="text/html;charset=UTF-8" language="java" %>
```
DV7Art, line 1: where your <%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %> ?

DV7Art: is it really work?

DV7Art (approved):
Nice work!

src/main/java/mate/controller/CreateCarController.java
```diff
@@ -0,0 +1,35 @@
...
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        req.getRequestDispatcher("/WEB-INF/cars/create.jsp").forward(req, resp);
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Car car = new Car(req.getParameter("model"),
+                manufacturerService.get(Long.valueOf(req.getParameter("manufacturerId"))));
```
rovein, line 31: Create a separate method to init car from request params.

Also, don't pass complex conditions to the constructor, instead try this:
```suggestion
        Car car = new Car(model, manufacturer);
```

src/main/java/mate/controller/CreateDriverController.java
rovein: Agree, let's make it consistent, use one naming strategy across all the controllers.

src/main/java/mate/controller/CreateDriverController.java
rovein: Yes, should be removed.

rovein (changes requested):
Good job! But let's improve your solution ;)

sarakhmen (approved):
Good job :)

PR #641, opened 2023-07-01
Tests done.

Hlib13:


src/main/java/mate/controller/cars/GetAllCarsController.java
```diff
@@ -0,0 +1,35 @@
...
+    private final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        List<Car> cars = carService.getAll();
+        req.setAttribute("cars", cars);
+        req.setAttribute("manufacturers", manufacturerService.getAll());
+        req.setAttribute("drivers", driverService.getAll());
```
okuzan, line 32: only cars are needed here

okuzan (changes requested):
cars package -> car. When you send redirect after adding, it makes more sense to navigate to the page where list of all entries is displayed, instead of to form for adding again

okuzan (changes requested):
**ignore this message**

src/main/java/mate/controller/cars/AddDriverToCarController.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest request, HttpServletResponse response)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(request.getParameter("driver_id"));
+        Long carId = Long.valueOf(request.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
+        response.sendRedirect(request.getContextPath() + "/index");
```
overpathz, line 36: Since we're working in the car context -> /cars

src/main/java/mate/controller/cars/CreateCarController.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturer_id"));
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturerService.get(manufacturerId));
+        car.setDrivers(new ArrayList<>());
```
overpathz, line 36: To get rid of explicitly setting an empty list you can init add an empty list within the constructor or directly in the field.

src/main/java/mate/controller/cars/CreateCarController.java
```diff
@@ -0,0 +1,40 @@
...
+        carService.create(car);
+        resp.sendRedirect(req.getContextPath() + "/index");
```
overpathz, line 38: After adding a car it would great to see them, so we have to redirect to /cars

overpathz (changes requested):

olekskov (approved):
Good job

PR #645, opened 2023-07-07
first commit

IhorPetrytskyi333:


src/main/java/mate/controller/AddDriverToCar.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.model.Driver;
+import mate.service.CarService;
+import mate.service.DriverService;
+
+@WebServlet(urlPatterns = "/drivers/addDriverToCar/add")
```
okuzan, line 15: image

okuzan (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/mate/controller/AddDriverToCarController.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver,car);
```
okuzan, line 38: let's redirect here to a page where we can see changes applied

okuzan (changes requested):
uncommit .ds_store files, separate controllers into packages

okuzan (changes requested):
please remove these files
image

PR #647, opened 2023-07-11
Controllers have been created. There is a successful operation page a…

RostyslavOnysh:
…nd a simple main page that showcases the entire functionality and brilliance of the design concept (which took half a lifetime to develop). The only thing I didn't quite understand was how to properly name the JSP files. Then the database was connected (seemingly successfully). In the model package, I added several constructors to the files located there. Then I tried to format everything correctly, but I got tangled up a thousand times, so I hope everything is working correctly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,6 +31,12 @@
             <version>4.0.1</version>
             <scope>provided</scope>
         </dependency>
+
+        <dependency>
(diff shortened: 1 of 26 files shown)


okuzan (approved):
Well done! 👏

PR #654, opened 2023-07-24
hm added

NSashchenko:


src/main/java/mate/controller/GetAllCarsController.java
```diff
@@ -0,0 +1,29 @@
...
+import mate.lib.Injector;
+import mate.model.Car;
+import mate.service.CarService;
+import mate.service.ManufacturerService;
+
+@WebServlet("/cars")
+public class GetAllCarsController extends HttpServlet {
+    private static final Injector injector = Injector.getInstance("mate");
+    private final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
```
olekskov, line 19: The variable 'manufacturerService' is declared but it is not used anywhere in the code. This is unnecessary and can be removed to make the code cleaner.

olekskov (approved):
Good job

PR #658, opened 2023-08-03
add my solution

mykhailoKrlKrk:


okuzan (changes requested):
Separate controllers into packages

olekskov (approved):
Good job

PR #662, opened 2023-08-06
realization jv-web-practice

bigvini:


src/main/java/mate/controller/car/AddCarController.java
```diff
@@ -0,0 +1,42 @@
...
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        String model = req.getParameter("model");
+        Manufacturer manufacturer = manufacturerService
+                .get(Long.valueOf(req.getParameter("manufacturer")));
+        Car car = new Car();
+        car.setModel(model);
+        car.setManufacturer(manufacturer);
+        car.setDrivers(new ArrayList<>());
+        carService.create(car);
```
okuzan, line 40: after post redirect to the page where you can see the changes just made (all cars after adding a car etc)

okuzan (changes requested):
Please fix the build before requesting another review

olekskov (approved):
Good job

PR #663, opened 2023-08-07
added controllers & appropriate jsp pages

TemkaHillfinger:


src/main/java/mate/controller/car/AddDriverToCarController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long driverId = Long.valueOf(req.getParameter("driver_id"));
+        Long carId = Long.valueOf(req.getParameter("car_id"));
+        Car car = carService.get(carId);
+        Driver driver = driverService.get(driverId);
+        carService.addDriverToCar(driver, car);
+
```
okuzan, line 34: add redirect to the page where you can see the change just made (all cars after adding a car etc)

okuzan (changes requested):

src/main/java/mate/controller/car/CreateCarController.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+            throws ServletException, IOException {
+        Long manufacturerId = Long.valueOf(req.getParameter("manufacturer_id"));
+        Car car = new Car();
+        car.setModel(req.getParameter("model"));
+        car.setManufacturer(manufacturerService.get(manufacturerId));
+        car.setDrivers(new ArrayList<>());
+        carService.create(car);
```
olekskov, line 34: Same as Oleh commented before. After all doPost method we should redirect user to page where he may see result of his request (list of all cars, drivers etc) otherwise he will struggle to understand if his request is successful.
Please recheck all controllers for same issue

olekskov (changes requested):

olekskov (approved):
Good job

